PR #4, opened 2019-06-28
Added a line for extended description for numpy style

paulkass:
# Prelude

By contributing to DoGe you agree to the following statements **(Put an `x` inside every [ ] you agree with)**:

# Why this PR?

According to the following link: https://docs.python-guide.org/writing/documentation/ there should be two sections with the numpy style guide. The first should be a summary
line, and then there should be a more detailed, extended description of
the function. This PR will add this extra TODO item for an extended
description.

### DIFF
--- a/ftplugin/python.vim
+++ b/ftplugin/python.vim
@@ -64,6 +64,8 @@ call add(b:doge_patterns, {
 \      'numpy': [
 \        '"""',
 \        'TODO',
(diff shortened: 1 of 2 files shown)


kkoomen:
I have read this as well, but from my experience, people tend to use 1 description (myself included), but sure, let's support it.

PR #7, opened 2019-07-15
Fix catch "Pattern not found"/"Unknown function" when locale is not en

KarolBedkowski:
# Why this PR?

The plugin does not work when vim locale is not en; i.e. for python files it raises an error on `<C-D>`:
```
Wykryto błąd podczas przetwarzania function doge#generate[3]..doge#generate#pattern:
wiersz   33:
E117: Nieznana funkcja: doge#preprocessors#python#tokens
Wykryto błąd podczas przetwarzania function doge#generate:
wiersz    3:
E171: Brak :endif 
```
 When calling not-existing functions, vim generates localized errors messages that do not match
catch statements in the plugin.

This patch replaces the current catching method for `E117` and `E486` by one method from `:h catch`: `catch /^Vim\%((\a\+)\)\=:<error number>/`

### DIFF
--- a/autoload/doge/generate.vim
+++ b/autoload/doge/generate.vim
@@ -38,7 +38,7 @@ function! doge#generate#pattern(pattern) abort
   try
     let l:preprocess_fn = printf('doge#preprocessors#%s#tokens', &filetype)
     call function(l:preprocess_fn)(l:tokens)
-  catch /E117: Unknown function/
+  catch /^Vim\%((\a\+)\)\=:E117/
   endtry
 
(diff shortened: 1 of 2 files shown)


kkoomen:
Hi!

Thank for the useful PR! Looks good and I should've used this instead yes, but it slipped my mind (._.)

kkoomen:
[v1.3.1](https://github.com/kkoomen/vim-doge/tree/v1.3.1) has been released within the master branch with your new fix in it. Thank you for your support!

PR #8, opened 2019-07-17
Add option to replace todo

vkasljevic:
# Prelude

# Why this PR?
This PR enables user to use configurable placeholder in generated documentation.
For ex. instead of using `TODO` as placeholder I could use `TODO(<user>)`.


kkoomen:
Thanks for putting quite some effort into this. However, I don't understand why you want to use `TODO(<user>)` to be set globally for everything? This would, for example, end up in:

```python
def myMethod(self: MyClass, p1: Sequence[T]) -> Generator[int, float, str]:
    """
    TODO(<user>)

    :param self MyClass: TODO(<user>)
    :param p1 Sequence[T]: TODO(<user>)
    :rtype Generator[int, float, str]: TODO(<user>)
    """
    pass
```

I'm not seeing in _why_ this would be useful.

vkasljevic:
Well, my case is that I have script for listing TODOs that matches only my `TODO(<user>)` because I work on large codebase with a lot of other people. I did this so that they could play along nicely, but I thought that someone else might have same inconvenience. Also, I don't have to keep my fork up to date.

As for the adding context to TODOs, it doesn't seem like that much extra work, I could add it if you want.

kkoomen:
I will add the context feature for now by myself, however I don't think this is something DoGe needs. If you want to give your descriptions more context for certain users you work with then this is a content issue and up to the user to add.

I'll close the issue for now. Thanks for submitting your idea.

kkoomen:
The TODO context support is released in [v1.4.0](https://github.com/kkoomen/vim-doge/tree/v1.4.0).

PR #22, opened 2019-07-19
update roxygen skeleton with return+examples

zkamvar:
# Prelude

# Why this PR?

This provide the return and examples fields, which are both good practice, necessary for package development, and are the defaults provided by RStudio's shortcut. 

By the way, I really like how this plugin "just works" ^\_^


kkoomen:
Hi,

Please make sure to follow the contribution guidelines and provide/update tests. If you want I can do the tests for you.

Can you more explain about `@examples`? I don't think this is needed when _not_ developing a package. I'm not familiar with R itself. I just learned the syntax of it so that DoGe can support it, but I don't know anything of the community and how people tend to write their code.

zkamvar:
> Please make sure to follow the contribution guidelines and provide/update tests. If you want I can do the tests for you.

Sorry about that! I'm a bit new to developing and testing vim plugins, and I accidentally updated the playground instead of the actual tests! 😅

I attempted to update the tests in a6b048b, but they still don't pass and I'm not quite sure how to load the development plugin to use the playground.

> Can you more explain about `@examples`? I don't think this is needed when _not_ developing a package. I'm not familiar with R itself. I just learned the syntax of it so that DoGe can support it, but I don't know anything of the community and how people tend to write their code.

Certainly. The examples is just an open field for adding quick examples of how the function should run. They aren't strictly necessary when not developing a package, but it is good practice to include at least one small example. Besides, it's what RStudio includes by default for their doc skeleton, so I figured, why not? ᕕ( ᐛ )ᕗ

kkoomen:
> I attempted to update the tests in a6b048b, but they still don't pass and I'm not quite sure how to load the development plugin to use the playground.

No worries. I'll fix the tests for you.

> Certainly. The examples is just an open field for adding quick examples of how the function should run. They aren't strictly necessary when not developing a package, but it is good practice to include at least one small example. Besides, it's what RStudio includes by default for their doc skeleton, so I figured, why not? ᕕ( ᐛ )ᕗ

Okay, I'll add it for now since you're ATM the only one requesting some changes for the R language.

kkoomen:
I've fixed the tests. You were almost there. I did change
```
"#' @examples",
"#' # TODO",
```
into
```
"#' @examples",
"#' TODO",
```
where I remove the additional `#` in front of the  `"#' # TODO",` into `"#' TODO",`. I assume this is not how it should be.

zkamvar:
Thank you for the help! 

Regarding the comment in front of the TODO: either way works!

Regarding the indentation of examples: it is not necessary. For example, [the ggplot2 package does not indent the examples](https://github.com/tidyverse/ggplot2/blob/d1ecd033c0a849b6bae16b07ad4b1e64b9a54109/R/aes.r#L37-L42):

```r
#' @examples
#' aes(x = mpg, y = wt)
#' aes(mpg, wt)
#'
#' # You can also map aesthetics to functions of variables
#' aes(x = mpg ^ 2, y = wt / cyl)
```

kkoomen:
Your contribution has been released as [v1.5.3](https://github.com/kkoomen/vim-doge/tree/v1.5.3).

Feel free to submit any new issues if you experience any unwanted behaviour in the future. Thanks for your contribution.

zkamvar:
Thank you for making this plugin with clear documentation and also for being patient!

PR #24, opened 2019-07-19
Fix python google style return type position

steven-lang:
The google style has the return type at the wrong position. See https://www.sphinx-doc.org/en/master/usage/extensions/example_google.html for examples.

# Prelude

# Why this PR?

[A clear and concise description of why this PR has been made]


steven-lang:
Argh, sorry, you just merged another commit in the same file. This PR has wrong changes now.

PR #32, opened 2019-07-28
Add the `:DogeGenerate` command to generate documentation

RobertAudi:
# Prelude

# Why this PR?

This just adds a command – `:DogeGenerate` – to generate documentation.

_**Note:** The `vimdoc` command did not add anything to the help file, so I took the initiative to add a section manually. I can remove it if necessary._

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ on a function, press `<Leader>d`, jump quickly through `TODO` items using
     + [`g:doge_mapping_comment_jump_backward`](#gdoge_mapping_comment_jump_backward)
     + [`g:doge_comment_interactive`](#gdoge_comment_interactive)
 - [Commands](#commands)
-  * [`:DogeGenerate`](#dogegenerate)
+    + [`:DogeGenerate`](#dogegenerate)
 - [Help](#help)
 - [FAQ](#faq)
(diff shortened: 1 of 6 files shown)


kkoomen:
Hi,

in [a47e59a](https://github.com/kkoomen/vim-doge/pull/32/commits/a47e59a8bea24375c69a57e92ae12131f98644d8) I did adjust the command declaration. The only thing that doesn't work with the `DogeGenerate` command is that the interactive mode doesn't seem to work for me. 

Apparently the [`l:todo_count`](https://github.com/kkoomen/vim-doge/blob/a47e59a8bea24375c69a57e92ae12131f98644d8/autoload/doge/generate.vim#L140) seems to be 0. It counts the number of `TODO` items in the comment. If there is 1 or more, it will jump to the first `TODO` item. For some reason it doesn't update the current buffer when doing a substitution. This has to do with my local vimrc since I've set `lazyredraw` in my local vimrc. I'll try to fix this.

kkoomen:
Sorry, using the `<Plug>` was the problem. I've removed this in the current commit and it's fixed now.

kkoomen:
This bug has been fixed and released in [v1.10.0](https://github.com/kkoomen/vim-doge/tree/v1.10.0).

PR #43, opened 2019-09-08
Use buffer mappings

mg979:
# Prelude

- [ x] I have ~read and understood~ currently no time for the [Contribution Guidelines](https://github.com/kkoomen/vim-doge/blob/master/CONTRIBUTING.md).
- [ x] I have ~read and understood~ currently no time for the [Code of Conduct](https://github.com/kkoomen/vim-doge/blob/master/CODE_OF_CONDUCT.md).

# Why this PR?

`<Tab>` and `<S-Tab>` are the most obvious mappings, but to dedicate them to
doge (and in all modes) is very demanding. Use buffer mappings instead,
so that they are only active in 'doge-mode', and unmapped when finished.

Deactivation is either automatic (end of placeholders, press outside of
comment area) or manual (main mapping is pressed again).

The first commit didn't give me troubles. The second commit requires some testing (I had a recursion once, then I changed it and it looks ok but some more testing is needed).

### DIFF
--- a/autoload/doge.vim
+++ b/autoload/doge.vim
@@ -5,13 +5,64 @@ set cpoptions&vim
 " @public
 " Generates documentation based on available patterns in b:doge_patterns.
 function! doge#generate() abort
+  if exists('b:doge_interactive')
+    return doge#deactivate()
+  endif
+  let success = 0
   if exists('b:doge_patterns')
     for l:pattern in get(b:, 'doge_patterns')
       if doge#generate#pattern(l:pattern) == v:false
         continue
+      else
+        let success = 1
       endif
-      return 1
+      if success
+        call doge#activate()
+      endif
+      return success
     endfor
+  endif
+endfunction
+
+""
+" @public
+" Activate doge mappings
+function! doge#activate() abort
+  let [f, b] = [g:doge_mapping_comment_jump_forward, g:doge_mapping_comment_jump_backward]
+  execute 'nmap <nowait><silent><buffer>' f '<Plug>(doge-comment-jump-forward)'
+  execute 'nmap <nowait><silent><buffer>' b '<Plug>(doge-comment-jump-backward)'
+  execute 'imap <nowait><silent><buffer>' f '<Plug>(doge-comment-jump-forward)'
+  execute 'imap <nowait><silent><buffer>' b '<Plug>(doge-comment-jump-backward)'
+  execute 'smap <nowait><silent><buffer>' f '<Plug>(doge-comment-jump-forward)'
+  execute 'smap <nowait><silent><buffer>' b '<Plug>(doge-comment-jump-backward)'
+  if get(g:, 'doge_activation_message', 1)
+    echo '[vim-doge] '
+    echohl Label
+    echon 'activated'
+    echohl None
+  endif
+endfunction
+
+""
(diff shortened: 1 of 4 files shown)


autoload/doge.vim
```diff
@@ -5,15 +5,66 @@ set cpoptions&vim
...
+  endif
+  let success = 0
   if exists('b:doge_patterns')
     for l:pattern in get(b:, 'doge_patterns')
       if doge#generate#pattern(l:pattern) == v:false
         continue
+      else
+        let success = 1
+      endif
+      if success
```
kkoomen: Use explicit check, such as: `if success == v:true`

autoload/doge.vim
```diff
@@ -5,15 +5,66 @@ set cpoptions&vim
 " @public
 " Generates documentation based on available patterns in b:doge_patterns.
 function! doge#generate() abort
+  if exists('b:doge_interactive')
```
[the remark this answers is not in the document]
mg979: In this branch it's working this way:

- `g:doge_comment_interactive` is false? mappings are not applied

Otherwise:

1. `\d` starts doge, and applies buffer mappings
2. `<Tab>/<S-Tab>` can cycle from last to first and viceversa, so you can start in the middle, press tab and you are brought to the ones you left behind, if not finished yet
3. `\d` again ? stop doge and unmap buffer mappings
4. whenever the last placeholder is reached, it stops and buffer mappings are removed
5. mappings are also removed if pressed outside of comment area

So the problem for you is that you don't want `\d` to disable mappings? The 'toggling' (3.) can be removed.

autoload/doge.vim
```diff
@@ -5,15 +5,69 @@ set cpoptions&vim
...
+  endif
+  let [f, b] = [g:doge_mapping_comment_jump_forward, g:doge_mapping_comment_jump_backward]
+  execute 'nmap <nowait><silent><buffer>' f '<Plug>(doge-comment-jump-forward)'
+  execute 'nmap <nowait><silent><buffer>' b '<Plug>(doge-comment-jump-backward)'
+  execute 'imap <nowait><silent><buffer>' f '<Plug>(doge-comment-jump-forward)'
+  execute 'imap <nowait><silent><buffer>' b '<Plug>(doge-comment-jump-backward)'
+  execute 'smap <nowait><silent><buffer>' f '<Plug>(doge-comment-jump-forward)'
+  execute 'smap <nowait><silent><buffer>' b '<Plug>(doge-comment-jump-backward)'
+  if get(g:, 'doge_activation_message', 0)
+    echo '[vim-doge] '
```
kkoomen: `[vim-doge]` should be `[DoGe]`.

kkoomen:
FYI: I will check the tests later this week.

mg979:
I rebased and made tests pass. Tests were probably failing because hitting `Tab` would cycle among placeholders also with default setup (no buffer mappings). I think it's a desirable behaviour, but tests are not made with it in mind. If you would like to make placeholder-cycling the default, one test should be fixed.

Right now every difference is optional and must be enabled with
```
let g:doge_buffer_mappings = 1
```

mg979:
This PR can be simplified a lot, I'm dumping it and making a new one.

PR #46, opened 2019-09-11
Ensure lazyredraw is enabled

mg979:
Enable lazyredraw in doge#activate(), if disabled.
Reenable in doge#deactivate() if enabled by DoGe.

# Prelude

# Why this PR?

[A clear and concise description of why this PR has been made]

### DIFF
--- a/autoload/doge.vim
+++ b/autoload/doge.vim
@@ -25,6 +25,12 @@ endfunction
 " @public
 " Activate doge buffer mappings, if option is set.
 function! doge#activate() abort
+  " Ensure lazyredraw is enabled
+  if &lazyredraw == v:false
+    set lazyredraw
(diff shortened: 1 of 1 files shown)


mg979:
It makes tests fail for some reason.

kkoomen:
It fails because I don't think [Vader](https://github.com/junegunn/vader.vim) is able to handle it. Unfortunately they have no issues about the lazyredraw in their issue tracker.

I think we should get rid of `lazyredraw` and maybe do something with `&hlsearch`. This is a better option and also for the tests.

Anyway, going to bed now. You can try to fix it in the meanwhile if possible.

mg979:
Thanks, this won't work so I'm closing it.

PR #49, opened 2019-09-11
Ensure lazyredraw is enabled (2)

mg979:
# Prelude

# Why this PR?

Similar to #46, but it is controlled by a global variable, that can be set to 0 to disable the check. I don't think it's important to document it because it can only be beneficial, at the same time the test vimrc must be able to disable it.

I'd prefer this approach to the discussed `hlsearch` one (that can also work), because if `hlsearch` isn't restored it can be annoying.

This can happen because right now mappings deactivation is not automatic, it waits for a `Tab` that can't find anything (as it was before, when there weren't buffer mappings).

`lazyredraw` is harmless in comparison and it's pointless not to use it because Vader apparently doesn't support it.

The second commit simply restores the old search register after deactivation.

### DIFF
--- a/autoload/doge.vim
+++ b/autoload/doge.vim
@@ -7,7 +7,7 @@ set cpoptions&vim
 function! doge#generate() abort
   let l:success = 0
 
-  " Store old search register
+  " Store old search register.
   let s:oldsearch = @/
 
   if exists('b:doge_patterns')
@@ -29,7 +29,7 @@ endfunction
 " @public
 " Activate doge buffer mappings, if option is set.
 function! doge#activate() abort
-  " Ensure lazyredraw is enabled
+  " Ensure lazyredraw is enabled.
   if get(g:, 'doge_lazyredraw', 1) && &lazyredraw == v:false
     set lazyredraw
     let s:doge_lazyredraw = 1


autoload/doge.vim
```diff
@@ -6,6 +6,10 @@ set cpoptions&vim
 " Generates documentation based on available patterns in b:doge_patterns.
 function! doge#generate() abort
   let l:success = 0
+
+  " Store old search register
+  let s:oldsearch = @/
```
kkoomen: What's the reason that this is not in the `doge#activate()` function?

autoload/doge.vim
```diff
@@ -44,8 +54,16 @@ endfunction
...
 function! doge#deactivate() abort
+  " Disable lazyredraw if it was previously enabled
+  if exists('s:doge_lazyredraw')
+    set nolazyredraw
+    unlet s:doge_lazyredraw
+  endif
   unlet b:doge_interactive
 
+  " Restore saved search register
+  let @/ = s:oldsearch
```
kkoomen: Shouldn't we put this inside `if exists('s:doge_lazyredraw')...endif`? We use this trick because of the `lazyredraw` issue, right?

kkoomen:
Besides the two comments: happy you came with another solution and I think it's a good option for now.

autoload/doge.vim
mg979: Because in `doge#activate()` the `TODO` pattern has already overwritten the old search pattern, it's triggered after the comment has been successfully added.

autoload/doge.vim
mg979: Sorry for the confusion.. the register saving/restoring has nothing to do with lazyredraw, it should have gone in another PR. It's meant to restore the search pattern after that it has been set to `TODO...` and mappings are deactivated.

PR #51, opened 2019-10-06
Deactivate doge if no placeholders left

mg979:
Check on InsertLeave if there are placeholders left, if not, deactivate
doge.

# Prelude

# Why this PR?

Before this PR, after the last placeholder was consumed, the first Tab keypress would still use doge mapping, which is undesiderable. It checks on InsertLeave, so that if the placeholder is reached, but not consumed, doge remains active.

### DIFF
--- a/autoload/doge/comment.vim
+++ b/autoload/doge/comment.vim
@@ -151,7 +151,6 @@ function! doge#comment#update_interactive_comment_info() abort
       let b:doge_interactive['lnum_comment_end_pos'] = l:lnum_comment_end_pos - 1
     endif
 
-
     " Update the amount of TODO items left.
     let b:doge_interactive['todo_count'] = doge#helpers#count(
           \ s:comment_placeholder,
--- a/autoload/doge/comment.vim
+++ b/autoload/doge/comment.vim
@@ -150,13 +150,6 @@ function! doge#comment#update_interactive_comment_info() abort
 
       let b:doge_interactive['lnum_comment_end_pos'] = l:lnum_comment_end_pos - 1
     endif
-
-    " Update the amount of TODO items left.
-    let b:doge_interactive['todo_count'] = doge#helpers#count(
-          \ s:comment_placeholder,
-          \ b:doge_interactive['lnum_comment_start_pos'],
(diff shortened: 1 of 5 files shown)


kkoomen:
I added proper documentation to the function, adjusted the function name to be more global and I did use the helpers function `doge#helpers#count()` to simplify the code. I moved your function code to the `TextChangedI` callback function and there it will properly update the new `todo_count` variable in `b:doge_interactive`. Your function will then only check if the amount of TODO items is `0` and will deactivate when the statement evaluates.

Any suggestions / recommendations or possible drawbacks towards my implementation I haven't thought about?

mg979:
Unfortunately this way TextChanged doesn't work, because if placeholders are deleted in normal mode that counter won't be updated (because it's updated when jumping). I tried a similar approach in my very first PR (checking the count) and I didn't like it. If you want it to work with TextChanged I don't think there are alternatives to search for the pattern in the autocommand, but it's not that there are performance issues or anything, so I don't see the problem.

kkoomen:
Agree, I haven't thought about that, but wasn't for of this while I implemented it. I changed the implementation by moving the update part in the `TextChangedI` function to your function. This works perfectly as I see for now. Agree?

mg979:
You sure it's working now? tabbing in insert mode seems broken to me.

kkoomen:
I do have the same. Didn't test insert-mode last time. This has to do with the `doge#helpers#count` function. It seems to be the `let l:cnt = execute(l:range . 's/' . a:word . '//gn')` line and I have no idea why. If I remove this and just `return 4` then it doesn't occur that the count returns `0`.

mg979:
I think you should learn a thing or two from [this guy](https://www.youtube.com/watch?v=seLT7M3OAgk&feature=youtu.be&t=393). Then maybe you can tell me what was wrong with the function I wrote :).

kkoomen:
I reverted back to your previous commit. The reason I prefer my version is because your code is redundant while there is a helper function that _should_ work, but the helper function is not being called at the auto-command`InsertLeave` so I'll just leave it at your version. That's fine for now.

PR #70, opened 2019-12-11
Add DogeCreateDocStandard (command)

mg979:
For custom template generation.


autoload/doge/pattern.vim
```diff
@@ -172,5 +172,101 @@ function! doge#pattern#generate(pattern) abort
...
+    endif
+  endfor
+  if has('nvim') && empty(l:path)
+    let l:path = stdpath('config')
+  endif
+  if !empty(l:path)
+    let l:path .= '/after/ftplugin/' . l:this_ft . '.vim'
+  endif
+  if filereadable(l:path)
+    exe 'split' fnameescape(l:path)
```
kkoomen: Since I use buffer tabs instead of windows, I suggest you open a tab if `&showtabline != v:false`, otherwise, use split.

kkoomen:
The command doesn't work for me. See demo.

Kapture 2019-12-12 at 10 07 04

autoload/doge/pattern.vim
mg979: Please let's keep it simple, `<c-w>T` can be used to move the buffer to a new tab if desired.

kkoomen: You can do `:enew` for a new buffer or `split` for a new buffer. That's a small adjustment I require to be supported.

mg979:
> 
> 
> The command doesn't work for me. See demo.
> 

Maybe you have `clipboard=unnamedplus` or something like that?
Can you try changing these lines?
```vim
  call setreg('"', l:doc)
  1
  normal! P'[=']
```
to
```vim
  call setreg('"', l:doc)
  1
  normal! ""P'[=']
```

mg979:
I fixed that cliboard problem and set it to use a new tab, I didn't understand that thing with `&showtabline`, you can change it to what you prefer.

kkoomen:
> Maybe you have clipboard=unnamedplus or something like that?

I have `set clipboard=unnamed`.

> I didn't understand that thing with &showtabline

See `:h showtabline`. If it's `1` or `2` then it means the user has enabled it. If it's `0` it means it's disabled, so you can simply do:

```vim
if &showtabline != v:false
  exe 'tabedit' fnameescape(l:path) 
else
  exe 'split' fnameescape(l:path) 
endif
```

kkoomen:
Pulling 293ac4c fixes the problem indeed. Can you implement the above code I mentioned?

kkoomen:
Are there any other things that needs to be done or need to be taken into consideration? If not, I'll write a test for this command soon.

mg979:
Is support for external programs for documentation generation implemented? If yes, the command could take it into account.

kkoomen:
> Is support for external programs for documentation generation implemented? If yes, the command could take it into account.

The only one we support now is `clang` for C-family languages. The only thing you have to adjust within the doc standard someone is extending is whether it has the `generator` key. If so, this will replace the `match` and `tokens` key in exchange for a `generator` key. You can see what I mean [here](https://github.com/kkoomen/vim-doge/blob/master/ftplugin/cpp.vim#L34).

kkoomen:
@mg979 I think that I will write tests for this and merge this. Agree?

mg979:
@kkoomen sorry I haven't time for this atm, the command is complete, if you need to make some adjustment you can make them by yourself.

PR #78, opened 2020-02-12
c/cpp: Use compile_commands if present to determine compiler args

d-karl:
The old behavior without a compile_commands.json present in the working
dir was kept available. If this file is present, the new mechanism
will use the actual compiler flags as specified by the build system to
parse the actual file the user is working in and retrieve info about the
token under the user's cursor.

# Prelude

- [x ] I have read and understood the [Contribution Guidelines](https://github.com/kkoomen/vim-doge/blob/master/CONTRIBUTING.md).
- [x ] I have read and understood the [Code of Conduct](https://github.com/kkoomen/vim-doge/blob/master/CODE_OF_CONDUCT.md).

# Why this PR?

In environments where the compiler generates a compile commands database (compile_commands.json). The clang toolchain can use this database to parse files in the exact way the compiler would. This enables reliable identification of the token type for which the user has requested documentation to be generated and associated tokens, such as parameters.
This PR is probably par from perfect, but it improves upon the existing mechanism if a compile commands database is present. The old mechanism for parsing has been preserved.

It should be noted that:
* It is assumed the compile_commands.json file is in the current working dir, as that should commonly be the project root and I didn't feel like searching for it, as that requires quite a bit of heuristics.
* The available token types are limited from DoGe's side.
* The current way of triggering the generator is suboptimal, since it is triggered 3 times, when one pass would be sufficient for the clang-based indexer. This makes documentation generation far slower than it would need to be, resulting in a noticable delay.
* Since header files have no entry inside the compile commands db, clang uses an internal heuristic to guess the correct flags for these files. This can very seldomly lead to incorrect results.

### DIFF
--- a/generators/libclang.py
+++ b/generators/libclang.py
@@ -31,8 +31,14 @@ If you compiled manually:
 Make sure that your $PATH and $LD_LIBRARY_PATH are set correctly.
 The libclang binary its location should be defined in the $LD_LIBRARY_PATH.
 """
-
-from clang.cindex import Index, CursorKind, Cursor
+from clang.cindex import (
+    Index,
+    CursorKind,
+    Cursor,
+    SourceLocation,
+    CompilationDatabase,
+    CompilationDatabaseError,
+)
 import sys
 import json
 import vim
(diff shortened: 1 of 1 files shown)


d-karl:
I just realized that I committed under the wrong author name. I will fix this soon, please hold off on merging for now.

generators/libclang.py
```diff
@@ -113,20 +119,80 @@ def print_func_decl(node: Cursor):
     :param node Cursor: The node to parse.
     """
     output = {}
-    output['name'] = node.spelling
-    output['returnType'] = node.result_type.spelling
-    output['parameters'] = []
+    output["name"] = node.spelling
```
kkoomen: Use single-quotes to match the project's standards.

generators/libclang.py
```diff
@@ -113,20 +119,80 @@ def print_func_decl(node: Cursor):
...
+                {"param-type": param_type, "name": child.spelling}
+            )
     print(json.dumps(output))
 
+
+def check_file_with_compile_flags(filename, args, filters, line, column):
+    """
+    Check the actual cpp file with the compile arguments retrieved from compile_commands.json.
+
+    :param filename: file to check
```
kkoomen: Capitalize the first letter of a description and add a period to the end of a description. The same goes for other comments.

kkoomen:
> I just realized that I committed under the wrong author name. I will fix this soon, please hold off on merging for now.

sure

<hr />

Thanks for your time for doing this. I really appreciate it. I haven't reviewed your code yet functionality-based, but it looks good enough. I'm not an experienced C dev, so any help I get is appreciated.

> The current way of triggering the generator is suboptimal, since it is triggered 3 times, when one pass would be sufficient for the clang-based indexer. This makes documentation generation far slower than it would need to be, resulting in a noticable delay.

This is true, since DoGe defines multiple patterns in the `ftplugin/c.vim`. Could you implement it in a way that the database compilations are cached, unless the last pattern in `ftplugin/c.vim` fails, or there is 1 pattern that matches?

Detecting the last patter requires a bit more work and I'll help you with it once we confirmed the caching strategy is an option.

kkoomen:
Closing due to new rewrite where clang is no longer needed, see #101.

PR #81, opened 2020-03-06
fix(ruby): Fix ruby keyword arguments and without parentheses

weirongxu:
# Prelude

# Why this PR?

Fix ruby function without parentheses

And I let `parameters.match` support the use of arrays because using one regular expression can't satisfy my needs.

And I let `l:curr_line_raw` contains the `\n` because ruby needs to determine if there is a `\n`.

https://github.com/weirongxu/vim-doge/blob/4c43cd41cf5b099b93225a3e2616f2544d639a6e/autoload/doge/pattern.vim#L29

### DIFF
--- a/autoload/doge/pattern.vim
+++ b/autoload/doge/pattern.vim
@@ -26,21 +26,8 @@ function! doge#pattern#generate(pattern) abort
     endif
   else
     " Skip if the current line does not match the main pattern.
-    let l:curr_line_raw = escape(doge#helpers#trim(join(l:lines, "\n")), '\')
-    let l:match = v:null
-    if type(a:pattern['match']) == type('')
-      if l:curr_line_raw =~# a:pattern['match']
-        let l:match = a:pattern['match']
-      endif
-    elseif type(a:pattern['match']) == type([])
-      for l:m in a:pattern['match']
-        if l:curr_line_raw =~# l:m
-          let l:match = l:m
-          break
-        endif
-      endfor
-    endif
-    if l:match is v:null
+    let l:curr_line_raw = escape(doge#helpers#trim(join(l:lines, ' ')), '\')
+    if l:curr_line_raw !~# a:pattern['match']
       return 0
     endif
 
@@ -56,7 +43,7 @@ function! doge#pattern#generate(pattern) abort
     " Extract the primary tokens.
     let l:tokens = get(doge#token#extract(
           \ l:curr_line,
-          \ l:match,
+          \ a:pattern['match'],
           \ a:pattern['tokens']
           \ ), 0, {})
   endif
--- a/autoload/doge/pattern.vim
+++ b/autoload/doge/pattern.vim
@@ -26,7 +26,9 @@ function! doge#pattern#generate(pattern) abort
     endif
   else
     " Skip if the current line does not match the main pattern.
-    let l:curr_line_raw = escape(doge#helpers#trim(join(l:lines, ' ')), '\')
+    let l:glue = has_key(a:pattern, 'normalize')
+          \ && a:pattern['normalize'] == v:false ? "\n" : ' '
+    let l:curr_line_raw = escape(doge#helpers#trim(join(l:lines, l:glue)), '\')
     if l:curr_line_raw !~# a:pattern['match']
       return 0
     endif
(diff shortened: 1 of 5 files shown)


kkoomen:
Hi.

Thanks for putting effort in the nice pull request, but you're making things more complicated then you need. The only solution you need is to allow ruby functions without parameters. The only thing you need to do is change [this](https://github.com/kkoomen/vim-doge/blob/master/ftplugin/ruby.vim#L45):

```rb
let s:function_and_class_method_pattern = doge#helpers#deepextend(s:pattern_base, {
\  'match': '\m^def\s\+\%([^=(!]\+\)[=!]\?\s*(\(.\{-}\))',
\  'tokens': ['parameters'],
\})
```

into this (notice the `match` regex has been adjusted)

```rb
let s:function_and_class_method_pattern = doge#helpers#deepextend(s:pattern_base, {
\  'match': '\m^def\s\+\%([^=(! ]\+\)[=!]\?\s*(\?\([^\n)]\+\))\?',
\  'tokens': ['parameters'],
\})
```

If you want to continue to do a PR then you can implement my fix and adjust tests to only this scenario. Otherwise I can fix it now if you want.

weirongxu:
```
let s:function_and_class_method_pattern = doge#helpers#deepextend(s:pattern_base, {
\  'match': '\m^def\s\+\%([^=(! ]\+\)[=!]\?\s*(\?\([^\n)]\+\))\?',
\  'tokens': ['parameters'],
\})
```
This patch will not completely fix the problem

```ruby
def myFunc format
end

def params(p1, p2)
end
```
Generate doc on `def myFunc` will get wrong results

kkoomen:
The following works:

```vim
let s:function_and_class_method_pattern = doge#helpers#deepextend(s:pattern_base, {
\  'match': '\m^def\s\+\%([^=(! ]\+\)[=!]\?\s*(\?\([^\n)]\{-}\))\?\s*end',
\  'tokens': ['parameters'],
\})

```

Any other scenarios?

weirongxu:
OK, but still not support this
```ruby
def myFunc format
  val = 1
end
```

weirongxu:
And another case to consider at the same time is that ruby parameters with parentheses support multiline.
```ruby
def myFunc(
  format,
  foo
)
end
```

kkoomen:
>OK, but still not support this
>
>def myFunc format
>  val = 1
>end

I've added a strict pattern for the parameter format. This works.

```vim
let s:function_and_class_method_pattern = doge#helpers#deepextend(s:pattern_base, {
\  'match': '\m^def\s\+\%([^=(! ]\+\)[=!]\?\s*(\?\s*\(\(\%(&\?[[:alnum:]_]\+\%(\s*=\s*[^, ]\+\)\?\)\%(\s*,\s*\)\?\)*\)\s*)\?.\{-}end',
\  'tokens': ['parameters'],
\})
```

Anything else?

weirongxu:
Thanks, it looks like could solve this issue, later I will add some tests for this.
Also, we should support the keyword parameter, https://thoughtbot.com/upcase/videos/ruby-keyword-arguments

weirongxu:
It seems necessary to check `\n`, otherwise there seems to be can't support following code
```ruby
def myFunc
end
```

kkoomen:
I allowed the use of [`'normalize': 0`](https://github.com/kkoomen/vim-doge/blob/5915891ca7ca578d05df2354db8ccc2fef12f5b7/ftplugin/ruby.vim#L48) in patterns which will not normalize the input (replacing `"\n"` with `' '`). Then I used `\n` as a check in the regex. This is a proper solution, since ruby is a big exception in it's language syntax.

All your tests are passing as well, thanks for adding them. If you don't have any other scenarios then I will merge soon.

weirongxu:
Thank you so much, you can merge now.

kkoomen:
This bug has been fixed and released in [v2.2.6](https://github.com/kkoomen/vim-doge/tree/v2.2.6). Thanks a lot for your contribution. It has been appreciated a lot. 🙌

<hr />

@weirongxu Are you interested in translating 2 README files into Chinese?

weirongxu:
> @weirongxu Are you interested in translating 2 README files into Chinese?

Sorry, no interest now, I need to do something else.

PR #94, opened 2020-08-30
WIP: Add support for async Python

### LINKED ISSUE #93: Python async functions
**Describe the solution you'd like**

Only sync functions are supported in Python. It would be nice to also include async functions in the detection algorithm.

eg:
```python
async def test(a: int) -> bool:
    pass
```

**Additional context**

I tried to update the regex myself but the \m at the start throws me off since it is not valid regex. 

```
\m(async )?def\s\+\%([^(]\+\)\s*(\(.\{-}\))\%(\s*->\s*\(.\{-}\)\)\?\s*:
```

Edit:

I'm currently reading the contribution guidelines, shouldn't be hard to implement these. Also, I now see the use of \m :sweat_smile:

LanderMoerkerke:
# Prelude

Todo:

- Tests
- Update playground
- Update PR

# Why this PR?

Support for async functions python, fixes #93

### DIFF
--- a/ftplugin/python.vim
+++ b/ftplugin/python.vim
@@ -56,7 +56,7 @@ let s:pattern_base = {
 " def myFunc(p1: Callable[[int], None] = False, p2: Callable[[int, Exception], None]) -> Sequence[T]:
 " ------------------------------------------------------------------------------
 let s:function_and_class_method_pattern = doge#helpers#deepextend(s:pattern_base, {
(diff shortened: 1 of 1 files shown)


LanderMoerkerke:
Test fail because the regex now sees "def" as a parameter. I only edited the function_and_class_method_pattern variable, to make the test pass, do I have to alter the base pattern or is my implementation of the function_and_class_method_pattern not valid?

kkoomen:
I appreciate your time and effort, but the regex are dizzy making. Just me fix this. Let's continue our discussion at #93.

LanderMoerkerke:
Thanks!

PR #104, opened 2020-10-11
fix(pattern): create custom doc only in valid rtp's

Sh3Rm4n:
# Prelude

By contributing to DoGe you agree to the following statements

# Why this PR?

This handles the special case, when .vim runtimepaths are lying around
on the system, but are not added in the runtimepath of the running
(neo)vim instances itself.

I had a `~/.vim` folder lying around in my system. 
Executing `:DogeCreateDocStandard ...` opened `~/.vim/after/ftplugin/...`, 
I use neovim and the `~/.vim` folder is no longer included in my runtimepath setup,
so I added this check.

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -16,7 +16,6 @@ cd $ROOT_DIR
 
 OS="$(uname)"
 OUTFILE="$ROOT_DIR/bin/vim-doge"
-
 PKG_VERSION=$(cat "$ROOT_DIR/.version")
 RELEASE_URL="https://github.com/kkoomen/vim-doge/releases/download/$PKG_VERSION"
 
@@ -29,13 +28,13 @@ else
   OUTFILE="$OUTFILE.exe"
 fi
 
-TEMP_FILE="vim-doge.tar.gz"
-DOWNLOAD_URL="$RELEASE_URL/$TARGET.tar.gz"
+FILENAME="$TARGET.tar.gz"
+DOWNLOAD_URL="$RELEASE_URL/$FILENAME"
 echo "Downloading $DOWNLOAD_URL"
 curl -L --progress-bar \
     --fail \
-    --output "$TEMP_FILE" \
+    --output "$FILENAME" \
     "$DOWNLOAD_URL"
-tar xzf "$TEMP_FILE" && mv "$TARGET" "$OUTFILE"
-rm -f "$TEMP_FILE"
+tar xzf "$FILENAME" && mv "vim-doge" "$OUTFILE"
+rm -f "$FILENAME"
 chmod +x "$OUTFILE"


autoload/doge/pattern.vim
```diff
@@ -185,6 +185,9 @@ function! doge#pattern#custom(name) abort
   let l:path = ''
   for l:p in ['~/.vim', '~/vimfiles']
     if isdirectory(expand(l:p))
+      if has('nvim') && match(&runtimepath, expand(l:p)) == -1
```
kkoomen, line 188: I guess we can just do it like this.

```
if isdirectory(expand(l:p)) && match(&runtimepath, expand(l:p)) >= 0
  " ...
endif
```

Sh3Rm4n: Yeah, that looks cleaner.

Sh3Rm4n:
Ok, I'm failing to adjust the test to pass. 

Optimally, this should fullfill following testcases

1. When `~.vim` exists and in runtimepath (which it is for vim)
2. When neovim-0.4.0, which has `stdpath()`
3. When no matching file is in the runtimepath (which is the case for nvim-0.2.0, as it does not have stdpath)

I'll look into that later today. 

Is it even possible to create directories in the Dockerfile?

kkoomen:
@Sh3Rm4n Just let me do the tests. I'm quick with it. I'll merge this later.

kkoomen:
> Optimally, this should fullfill following testcases
> 
>     1. When `~.vim` exists and in runtimepath (which it is for vim)

`~/.vim` also exists for neovim so I assume no need to change this as this is already being checked in the current test, right?
 
>     2. When neovim-0.4.0, which has `stdpath()`

What exactly do you want to test here?

>     3. When no matching file is in the runtimepath (which is the case for nvim-0.2.0, as it does not have stdpath)

When no matching file is in the runpath ... then what?

Sh3Rm4n:
Sorry for not being clear about what I want to test. 
First and foremost I want the current test to pass with this change. 

> ~/.vim also exists for neovim so I assume no need to change this as this is already being checked in the current test, right?

With the `isdirectory` check, a buffer with `~/.vim/after/ftplugin/<...>.vim` as path is no longer created, 
if this path is not in `runtimepath`. This is the case for neovim, as it does not include `.vim` in the `runtimepath` as I check with `nvim -u NONE +"echo &rtp"`.

So as this conditional is no longer executed:

https://github.com/kkoomen/vim-doge/blob/9604d8b28a0d48da98abd960471ec4d330059ce4/autoload/doge/pattern.vim#L187

it will hit this conditional

https://github.com/kkoomen/vim-doge/blob/9604d8b28a0d48da98abd960471ec4d330059ce4/autoload/doge/pattern.vim#L192

And for `nvim-0.2.x`, as neovim got the `stdpath` function in neovim 0.3.0, it will hit this path:

https://github.com/kkoomen/vim-doge/blob/9604d8b28a0d48da98abd960471ec4d330059ce4/autoload/doge/pattern.vim#L205

> When no matching file is in the runpath ... then what?

But it could be the case, that there is any (n)vim session with no matching runtimepath, e.g. assuming `.vim` and `~/vimfiles` are removed from the rtp of a vim instance.  

In conclusion, with this change we are forced to test all these paths, so that the CI is passing for vim7, vim8, nvim-0.2 and nvim-0.4.

I hope I made it a bit more clear now.

kkoomen:
@Sh3Rm4n Well, since you seem to know the ins and outs, I suggest you write the tests. Mention me when you're done and I'll review it again.

Sh3Rm4n:
Before I go about writing the tests, is it possible to create directories or files in the CI container?

https://travis-ci.com/github/kkoomen/vim-doge/jobs/397935105#L268

This is where I gave up, because I do not know how to test it otherwise.

kkoomen:
@Sh3Rm4n No need to create those directories . When we create a doc standard we open a new file with a given path but don't save it. The user can choose to save it. If those directories don't exists then the user should create it first, vim-doge is not responsible whether those directories exists or not, hence why we check the current buffer's path via `expand('%:p')`. 

You can remove all the `call mkdir()` and `execute "!rmdir"` statements.

Also, use more modern syntax such as using `execute()` as a function instead of `execute` as a command and pay attention to the repository's standards by adding a single white-space before and after a dot operator `.`.

Sh3Rm4n:
Should be ready for review.

kkoomen:
@Sh3Rm4n Sorry for force-pushing and screwing up the PR. I can't (force) push to your fork. Can you recreate the fork based off the new master branch and then download the following `vim-doge.patch` to re-add your functionality:

<details>
<summary>vim-doge.patch</summary>

```
diff --git a/autoload/doge/pattern.vim b/autoload/doge/pattern.vim
index 510b0f5..282590b 100644
--- a/autoload/doge/pattern.vim
+++ b/autoload/doge/pattern.vim
@@ -184,17 +184,20 @@ function! doge#pattern#custom(name) abort
   " otherwise create a new file with an appropriate path.
   let l:path = ''
   for l:p in ['~/.vim', '~/vimfiles']
-    if isdirectory(expand(l:p))
-      let l:path = expand(l:p)
+    let l:p = expand(l:p)
+    if match(&runtimepath, l:p) >= 0
+      let l:path = l:p
       break
     endif
   endfor
-  if has('nvim') && empty(l:path)
-    let l:path = stdpath('config')
-  endif
-  if !empty(l:path)
-    let l:path .= '/after/ftplugin/' . l:this_ft . '.vim'
+  if empty(l:path)
+    if exists('*stdpath')
+      let l:path = stdpath('config')
+    else
+      let l:path = getcwd()
+    endif
   endif
+  let l:path .= '/after/ftplugin/' . l:this_ft . '.vim'
   if filereadable(l:path)
     let l:cmd = &showtabline ? 'tabedit' : 'split'
     call execute(l:cmd . fnameescape(l:path), 'silent!')
@@ -204,9 +207,7 @@ function! doge#pattern#custom(name) abort
   else
     call execute(&showtabline ? 'tabnew' : 'new', 'silent!')
     setfiletype vim
-    if !empty(l:path)
-      call execute('file ' . fnameescape(l:path), 'silent!')
-    endif
+    call execute('file ' . fnameescape(l:path), 'silent!')
   endif
 
   " Generate the template and paste it at the top of the file.
diff --git a/test/commands/create-doc-standard.vader b/test/commands/create-doc-standard.vader
index 0b8cf12..263210a 100644
--- a/test/commands/create-doc-standard.vader
+++ b/test/commands/create-doc-standard.vader
@@ -5,6 +5,9 @@
 # ==============================================================================
 
 Execute (Create a new doc standard based on phpdoc, setup dummy pattern and execute the DogeCreateDocStandard command):
+  if !has('nvim')
+    let &runtimepath .= ',' . expand('~/.vim')
+  endif
   enew
   setfiletype php
   let b:doge_patterns = {
@@ -48,7 +51,15 @@ Execute (Create a new doc standard based on phpdoc, setup dummy pattern and exec
   DogeCreateDocStandard phpdoc
 
 Then (Expect a file to be created with specific content):
-  AssertEqual expand('~/.vim/after/ftplugin/php.vim'), expand('%:p')
+  let b:path = '/after/ftplugin/php.vim'
+  if !has('nvim')
+    let b:path = '~/.vim' . b:path
+  elseif exists('*stdpath')
+    let b:path = '~/.config/nvim' . b:path
+  else
+    let b:path = getcwd() . b:path
+  endif
+  AssertEqual expand(b:path), expand('%:p')
 
   let b:content = join([
   \  '" Preserve existing doge settings.',
@@ -81,16 +92,27 @@ Then (Expect a file to be created with specific content):
   AssertEqual b:content, join(getline(line('^'), line('$')), "\n")
   unlet b:content
 # Remove the buffer to ensure every test will create a new doc standard file.
-  call delete(expand('~/.vim/after/ftplugin/php.vim'))
+  call delete(expand(b:path))
   bdelete!
 
 Execute (Create a new doc standard for an unsupported filetype, setup dummy pattern and execute the DogeCreateDocStandard command):
+  if !has('nvim')
+    let &runtimepath .= ',' . expand('~/.vim')
+  endif
   enew
   setfiletype foo
   DogeCreateDocStandard foo
 
 Then (Expect a file to be created with specific content):
-  AssertEqual expand('~/.vim/after/ftplugin/foo.vim'), expand('%:p')
+  let b:path = '/after/ftplugin/foo.vim'
+  if !has('nvim')
+    let b:path = '~/.vim' . b:path
+  elseif exists('*stdpath')
+    let b:path = '~/.config/nvim' . b:path
+  else
+    let b:path = getcwd() . b:path
+  endif
+  AssertEqual expand(b:path), expand('%:p')
 
   let b:content = join([
   \  '" Preserve existing doge settings.',
@@ -123,5 +145,5 @@ Then (Expect a file to be created with specific content):
   AssertEqual b:content, join(getline(line('^'), line('$')), "\n")
   unlet b:content
 # Remove the buffer to ensure every test will create a new doc standard file.
-  call delete(expand('~/.vim/after/ftplugin/foo.vim'))
+  call delete(expand(b:path))
   bdelete!
```
</details>

Then after downloading it, run: `patch -p1 < vim-doge.patch`, commit it and re-open another PR.

PR #108, opened 2020-10-12
Windows support for install script

bartlibert:
# Prelude

# Why this PR?

Installing tree-sitter using the "install.sh" script only works on linux, this PR adds a windows-compatible "install.bat" script and calls that one from the "doge#install()" function if it detects vim is running on a windows system.

### DIFF
--- a/autoload/doge.vim
+++ b/autoload/doge.vim
@@ -203,6 +203,12 @@ function! doge#install() abort
   let l:args = !empty(l:langs) ? l:langs : l:languages
   echo '[DoGe] Installing parsers: ' . join(l:args, ', ')
   call execute('!' . g:doge_dir . '/install.sh ' . join(l:args, ' '))
+  if has('win32')
+    let l:install_script = 'install.bat'
+  else
+    let l:install_script = 'install.sh'
+  endif
+  execute('!' . g:doge_dir . '/' . l:install_script . ' ' . join(l:args, ' '))
 endfunction
 
 let &cpoptions = s:save_cpo
--- a/install.bat
+++ b/install.bat
@@ -1,0 +1,27 @@
+@echo off
+set ROOT_DIR=%~dp0
+cd %ROOT_DIR%
+
+set packages=
+
+:gen_package_names
+if [%1]==[] goto package_names_ready
+  set package_name=tree-sitter-%1
+  @echo Preparing to install package: %package_name%
+  set packages=%packages% %package_name%
+  shift
+goto gen_package_names
+
+:package_names_ready
+if exist .\node_modules\ goto install_packages
+  call npm i --only=production --no-save
+
+:install_packages
+if "%packages%"=="" goto run_build
+  call npm i --no-save %packages%
+
+:run_build
+if exist .\dist\index.js goto done
+  call npm run build
+
+:done


kkoomen:
Hi. Thanks for putting effort into the Windows compatibility. Today I've merged #111 so this does PR does require changes based on #111. I don't use Windows myself, so if you can add support for this then that'd be great.

Please have a look at the new `scripts/install.sh` in master. Right now we do a check based on the `uname` and if it doesn't match `Darwin` or `Linux` then we assume windows. I rather remove this part and use a separate `install.bat` just like you did in this PR.

bartlibert:
I'll have a look, but it will probably be later this week. I'm also not a windows expert (just forced to use it at work), so it'll take some time to figure out the best way, as stuff like curl is not (by default) available on windows.

kkoomen:
@bartlibert Thanks for your effort. If you have a windows alternative this friday, then I'll use that. Otherwise I'd like to use Powershell, since it does have built-in support for http requests, see [this](https://superuser.com/questions/299754/wget-curl-alternative-native-to-windows/494003#494003) stackoverflow answer.

bartlibert:
Hmm, using powershell from vim might be a bit more finicky since vim does
not support it natively, but it might be the way to go unfortunately...

Op ma 19 okt. 2020 om 14:05 schreef Kim 金可明 <notifications@github.com>:

bartlibert:
Well, it's a bit more complicated then "out of the box support".
By default, on windows, vim uses `cmd.exe` for its shell. This means that executing `:!install.bat` will execute `install.bat` in the `cmd.exe` program. `.bat` files are also supported in powershell, so even if the user has set powershell as its vim shell, it will still work.

The issue with using powershell (trust me, I'd prefer to use it as well), is that you have to versions, powershell (which uses `powershell.exe` as its executable) and powershell core (`pwsh.exe` as its executable) which are not fully compatible. Also, different versions of powershell core support different functions, so it'll take some work to make everything robust.

kkoomen:
@bartlibert Clear. Well, you clearly know more than me, so I take your vote. Do what you think is best.

P.s.: I just updated the install.sh a bit, since the new version does use tarballs which decreases the overall binary assets size by a factor of 3, so this should be handled as well. Have a look later ;)

kkoomen:
@bartlibert Please sync your fork with the original master by pulling from kkoomen/vim-doge, because I've updated the CI which are also crucial for your PR.

kkoomen:
@bartlibert You can leave it up to me. I'll add support for Windows through powershell because there are more scripts and things that have to be created.

I'll close your issue for now. Many thanks for your time. If something is not working later on Windows, feel free to submit a PR.

If Windows support is there I'll comment on this issue to give you a headsup.

bartlibert:
Ok thanks, I have not had time to look at it in detail yet, due to work and
family commitments. One thing I thought about is to use the `executable`
function in vimscript to check if `pwsh.exe` or `powershell.exe` is
present, to be able to use the correct interpreter (in 99% of cases I would
guess pwsh.exe has priority over powershell.exe if both are present). If
none are present, some manual instructions might be shown (or a link to
github or something similar)

Op ma 26 okt. 2020 om 15:22 schreef Kim 金可明 <notifications@github.com>:

kkoomen:
@bartlibert I'll take it with me when implementing it. Thanks for the suggestion.

kkoomen:
@bartlibert I just released [v3.5.1](https://github.com/kkoomen/vim-doge/tree/v3.5.1) with proper support for Windows. Have a look, test some scenario's and let me know what you think and if you have improvements. The tests are passing and I've tested myself as well on a virtual machine.

kkoomen:
@bartlibert Also, let me know if you are using Vim or NeoVim. I tested both and both seem okay.

bartlibert:
I've tested it with the latest vim nightly on windows and it seems to work
fine. Maybe one thing to consider: while "regular" windows powershell is
and will be a core part of Windows for the foreseeable future, microsoft is
actively promoting the use of "powershell core" (even showing a message
about it when starting windows powershell), so for now it's not an issue,
but it might get deprecated in the future, so keep in mind that you'll have
to use "pwsh.exe" instead of "powershell.exe" then. (The new install script
works in powershell core, so that's fine already ;) )

Op zo 1 nov. 2020 om 04:48 schreef Kim 金可明 <notifications@github.com>:

kkoomen:
@bartlibert Thanks for testing! I'll try to use pwsh.exe as a first option with a fallback to powershell.exe later this week.

kkoomen:
@bartlibert Actually, do you think I can just use pwsh.exe and ignore powershell.exe?

bartlibert:
Well, you can't be sure everybody has installed powershell core (pwsh.exe),
as this is an optional program, while windows powershell (powershell.exe)
is (for the time being) always installed on Windows, so better to support
both I guess.

Op di 3 nov. 2020 om 12:17 schreef Kim 金可明 <notifications@github.com>:

kkoomen:
@bartlibert For now I'll just keep powershell.exe. If someone faces a problem then I welcome them to create an issue ;)

PR #153, opened 2020-11-10
Better `doge#install()`

subnut:
# Prelude

# Why this PR?

Please see the detailed commit messages

### DIFF
--- a/autoload/doge.vim
+++ b/autoload/doge.vim
@@ -181,15 +181,15 @@ endfunction
 "" @public
 " Install the necessary dependencies.
 function! doge#install() abort
-  func! s:report_result(exitcode) abort
+  function! s:report_result(exitcode) abort
     if a:exitcode == 0
-      echom 'vim-doge installed sucessfully'
+      echom '[DoGe] installed sucessfully'
     else
       echohl ErrorMsg
-      echom 'vim-doge could not be installed ' . '(exit code: ' . a:exitcode . ')'
+      echom '[DoGe] installation failed ' . '(exit code: ' . a:exitcode . ')'
       echohl None
     endif
-  endfun
+  endfunction
 
   if has('win32')
     let l:command = (executable('pwsh.exe') ? 'pwsh.exe' : 'powershell.exe')
@@ -202,9 +202,11 @@ function! doge#install() abort
 
   if has('nvim') && exists(':terminal') == 2
     " neovim with :terminal support
+    " vint: next-line -ProhibitUnusedVariable
     function! s:callback(...) abort
-      if a:2 == 0   " exitcode
-        call execute(s:terminal_bufnr . 'bd!')         " no errors to report, so delete buffer
+      if a:2 == 0 " exitcode
+        " no errors to report, so delete buffer
+        call execute(s:terminal_bufnr . 'bd!')
         unlet s:terminal_bufnr
       endif
       call s:report_result(a:2)
@@ -213,19 +215,20 @@ function! doge#install() abort
     enew
     call termopen(l:command, {'on_exit': function('s:callback')})
     let s:terminal_bufnr = bufnr()
-
   elseif has('nvim')
-    " neovim does not output stdout if called using :call execute()
-    " to show download progress bar, directly run execute()
+    " Neovim does not show any stdout output if called with :call execute(),
+    " therefore to show the download progress bar, we need to call execute() by
+    " itself.
     execute('!' . l:command)
     call s:report_result(v:shell_error)
-
   elseif has('terminal')
     " vim with +terminal
+    " vint: next-line -ProhibitUnusedVariable
     function! s:callback(channel) abort
       let l:exitcode = job_info(ch_getjob(a:channel)).exitval
       if l:exitcode == 0
-        call execute(s:terminal_bufnr . 'bd!')         " no errors to report, so delete buffer
+      " no errors to report, so delete buffer
+        call execute(s:terminal_bufnr . 'bd!')
         unlet s:terminal_bufnr
       endif
       call s:report_result(l:exitcode)
--- a/autoload/doge.vim
+++ b/autoload/doge.vim
@@ -182,9 +182,27 @@ endfunction
 " Install the necessary dependencies.
 function! doge#install() abort
   if has('win32')
-    call execute('!powershell.exe ' . g:doge_dir . '/scripts/install.ps1')
+    let l:command = 'powershell.exe Set-ExecutionPolicy -ExecutionPolicy Unrestricted -Scope Process -Force; ' . g:doge_dir . '/scripts/install.ps1'
   else
-    call execute('!' . g:doge_dir . '/scripts/install.sh')
+    let l:command = g:doge_dir . '/scripts/install.sh'
+  endif
+  if has('nvim') && exists(':terminal') == 2
+    " neovim with :terminal support
+    function! Callback(...) abort
+      if a:2 == 0   " exitcode
+        bd!         " no errors to report, so delete buffer
+      endif
+    endfun
+    sp
+    enew
+    call termopen(l:command, {'on_exit': function('Callback')})
+  elseif has('nvim')
+    " neovim does not output stdout if called using :call execute()
+    " to show download progress bar, directly run execute()
+    execute('!' . l:command)
+  else
+    " vim
+    call execute('!' . l:command)
   endif
 endfunction
 
--- a/autoload/doge.vim
+++ b/autoload/doge.vim
@@ -181,7 +181,7 @@ endfunction
 "" @public
 " Install the necessary dependencies.
 function! doge#install() abort
-  func! Report(exitcode) abort
+  func! s:report_result(exitcode) abort
     if a:exitcode == 0
       echom 'vim-doge installed sucessfully'
     else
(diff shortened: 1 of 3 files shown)


subnut:
Ping @kkoomen

kkoomen:
Thanks for your effort and +1 for the nice descriptions in the comments. I'll test it this weekend myself.

kkoomen:
@subnut Since you did fix something for Windows, do you know if it's possible to check whether the user has `pwsh.exe`, if so then use that, otherwise do a fallback to `powershell.exe`. If you have any idea that feel free to submit another PR.

subnut:
@kkoomen Yep, it's easy. I am implementing it right away ... but can you please explain what is the advantage of `pwsh.exe` over the default `powershell.exe` ? 
I am actually not that familiar with PowerShell Core ....

kkoomen:
See [this comment](https://github.com/kkoomen/vim-doge/pull/108#issuecomment-721010001) in a previous discussion.

subnut:
@kkoomen Done!

autoload/doge.vim
```diff
@@ -181,10 +181,62 @@ endfunction
...
+    let l:term_height = 8
+  else
+    let l:command = fnameescape(g:doge_dir) . '/scripts/install.sh'
+    let l:term_height = 4
+  endif
+
+  if has('nvim') && exists(':terminal') == 2
+    " neovim with :terminal support
+    function! s:callback(...) abort
+      if a:2 == 0   " exitcode
```
kkoomen, line 206: Please remove unnecessary spaces:

```
 if a:2 == 0   " exitcode 
```

to

```
if a:2 == 0 " exitcode 
```

autoload/doge.vim
```diff
@@ -181,10 +181,62 @@ endfunction
...
+        call execute(s:terminal_bufnr . 'bd!')         " no errors to report, so delete buffer
```
kkoomen, line 207: I suggest to put this comment above the `call execute()` statement.

autoload/doge.vim
```diff
@@ -181,10 +181,62 @@ endfunction
...
+        unlet s:terminal_bufnr
+      endif
+      call s:report_result(a:2)
+    endfun
+    exe (&splitbelow ? 'botright' : 'topleft') . l:term_height . 'sp'
+    enew
+    call termopen(l:command, {'on_exit': function('s:callback')})
+    let s:terminal_bufnr = bufnr()
+
```
kkoomen, line 216: This empty line can be deleted.

autoload/doge.vim
```diff
@@ -181,10 +181,62 @@ endfunction
...
+    " to show download progress bar, directly run execute()
+    execute('!' . l:command)
+    call s:report_result(v:shell_error)
+
+  elseif has('terminal')
+    " vim with +terminal
+    function! s:callback(channel) abort
+      let l:exitcode = job_info(ch_getjob(a:channel)).exitval
+      if l:exitcode == 0
+        call execute(s:terminal_bufnr . 'bd!')         " no errors to report, so delete buffer
```
kkoomen, line 228: I suggest to put this comment above the `call execute()` statement.

autoload/doge.vim
```diff
@@ -181,10 +181,62 @@ endfunction
...
+    enew
+    call termopen(l:command, {'on_exit': function('s:callback')})
+    let s:terminal_bufnr = bufnr()
+
+  elseif has('nvim')
+    " neovim does not output stdout if called using :call execute()
+    " to show download progress bar, directly run execute()
+    execute('!' . l:command)
+    call s:report_result(v:shell_error)
+
```
kkoomen, line 222: This empty line can be deleted.

scripts/install.sh
```diff
@@ -1,17 +1,26 @@
...
 # https://www.gnu.org/software/bash/manual/bash.html#The-Set-Builtin
 set -e
 set -u
 
+if which curl &> /dev/null
+then
+  continue
+else
+  echo "curl: command not found" >&2
+  echo "Please ensure curl is installed on the system and is available in the PATH" >&2
```
kkoomen, line 13: I suggest to use the following message: `Please make sure that curl is installed and available in your $PATH`

autoload/doge.vim
```diff
@@ -181,10 +181,62 @@ endfunction
 "" @public
 " Install the necessary dependencies.
 function! doge#install() abort
+  func! s:report_result(exitcode) abort
+    if a:exitcode == 0
+      echom 'vim-doge installed sucessfully'
```
kkoomen, line 186: Message standards from the plugin are always prefixed with `[DoGe]`, to follow this standard I suggest you use `echom '[DoGe] installed successfully'`

autoload/doge.vim
```diff
@@ -181,10 +181,62 @@ endfunction
...
+    else
+      echohl ErrorMsg
+      echom 'vim-doge could not be installed ' . '(exit code: ' . a:exitcode . ')'
```
kkoomen, line 189: Message standards from the plugin are always prefixed with `[DoGe]`, to follow this standard I suggest you use `echom '[DoGe] installation failed: .....'`

autoload/doge.vim
```diff
@@ -181,10 +181,62 @@ endfunction
 "" @public
 " Install the necessary dependencies.
 function! doge#install() abort
+  func! s:report_result(exitcode) abort
```
kkoomen, line 184: Please use `function!` and `endfunction` instead of abbreviations to abide by the plugin's standards.

autoload/doge.vim
```diff
@@ -181,10 +181,62 @@ endfunction
...
+      call s:report_result(a:2)
+    endfun
+    exe (&splitbelow ? 'botright' : 'topleft') . l:term_height . 'sp'
+    enew
+    call termopen(l:command, {'on_exit': function('s:callback')})
+    let s:terminal_bufnr = bufnr()
+
+  elseif has('nvim')
+    " neovim does not output stdout if called using :call execute()
+    " to show download progress bar, directly run execute()
```
kkoomen, line 219: Since this is quite a long explanation it's recommended to use a capital letter and dot at the end of the sentence. I suggest you use something like:

```
" Neovim does not output any stdout if it's called with :call execute(), therefore to show the download progress bar, we just call execute() by itself.
```

**NOTE**: Make sure to wrap at the 80th column.

kkoomen (approved):
I judged your code just on logic and styleguide. I haven't tested it yet, but I will do that this weekend somewhere.

kkoomen (changes requested):
I submitted the previous wrong, as I do require changes.

scripts/install.sh
subnut: Why `make sure` instead of `ensure`?
In my opinion, both convey the same meaning, but the latter seems more proper

kkoomen: Just feel more native and a bit more colloquial.

scripts/install.sh
```diff
@@ -1,17 +1,26 @@
...
 # https://www.gnu.org/software/bash/manual/bash.html#The-Set-Builtin
 set -e
 set -u
 
+if which curl &> /dev/null
+then
+  continue
+else
+  echo "curl: command not found" >&2
+  echo 'Please make sure curl is installed on the system and is available in the $PATH' >&2
```
kkoomen, line 13: You didn't copy my suggestion from before. I did remove `on the system` because it sounds weird and is redundant. Please remove it from the message.

kkoomen (changes requested):

scripts/install.sh
subnut: Oops.. sorry!

subnut: Fixed.

kkoomen:
I just tested it but this doesn't work comfortably with tabs. You can add a check with `if &showtabline != 0 `. The problem is that I first run `:call doge#install()` and then while it's downloading, I am switching buffers. After that the stdout will appear double. See demo below

Also, the Linter failed in your last commit. See [details](https://github.com/kkoomen/vim-doge/pull/153/checks?check_run_id=1396007751).

vim-doge-install

subnut:
Hmm... I can't reproduce your problem. Can you please give me a minimal .vimrc?

kkoomen:
Use this minimal vimrc:

```vim
set nocompatible
set runtimepath^=/path/to/vim-doge
filetype plugin indent on
syntax on
set hidden
set showtabline=2
```

then `vim -u vim-doge.vimrc` and then do the following:
- run `:call doge#install()`
- presss <kbd>Ctrl</kbd> + <kbd>w</kbd> + <kbd>ArrowDown</kbd>
- run `:bprev`

subnut:
Thanks, those repro steps helped a lot!

Should be fixed in the latest commit. Please check.

kkoomen:
Tested it. Works really well! How does this work on Windows? I don't run Windows myself. Can you send a GIF or another type of visual demo to show this behavior?

subnut:
Here you go -

gifcast_201113210104

kkoomen:
That looks clean. I like it. One thing, if possible, can you add some kind of text while it's preparing the download? So when you run `:call doge#install()` it's preparing the download and basically shows you an empty terminal for roughly 5 seconds. Can we output some text like: `Preparing to download vim-doge binary...` ?

subnut:
Like so .... ? (See GIF)
gifcast_201113211038

kkoomen:
Perfect! If you push now then I'll merge tomorrow morning, since it's late here.

kkoomen (approved):

kkoomen:
This feature will be released in the upcoming version [v3.6.0](https://github.com/kkoomen/vim-doge/tree/v3.6.0).

subnut:
> I can tell that your git commit hooks didn't trigger due to this.

I think there are no hooks...
image

kkoomen:
Just make sure you run `npm install` before developing.

Issue #68, opened 2019-12-04
Avoid type redundancy in TypeScript


resolritter:
**Is your feature request related to a problem? Please describe.**

When generating documentation with JSDoc for TypeScript, it's redundant to add the type to the comments because they are already specified in the functions' signatures. In the editor, the IDE integration tells us so:

![image](https://user-images.githubusercontent.com/17429390/70141424-cdebf880-1675-11ea-8697-265f457579a1.png)

Aside from redundancy, another issue is that if we change the function's types, the JSDoc will be outdated with the old types.

**Describe the solution you'd like**

One of two solutions:

1. If we omit the types in JSDoc, this is no longer an issue and the warning goes away. I think I prefer this because TSDoc is still a work-in-progress and the spec is not fully mature yet, it seems.

2. The alternative, TSDoc, does not include the types into the comments. However, the spec is under discussion and it will surely be constantly changing in the near future, so I think it's not the time to add support for it.
https://github.com/microsoft/tsdoc

**Describe alternatives you've considered**

Manually removing each type from the comments isn't good.

kkoomen:
I do agree with the redundancy. For me it would be more logical to add the type if the type is not listed in the parameter itself, otherwise we can remove it. Would that be an idea?

Regarding the TSDoc, love to support it but as you said, it's heavily discussed so it's better to wait for a while now.

resolritter:
Your idea is even better than what I was imagining @kkoomen !

If it's not too much trouble, you could omit those type annotations only through a configuration variable, leaving the default as it is currently. That warning is coming from [tsserver](https://github.com/Microsoft/TypeScript/wiki/Standalone-Server-%28tsserver%29) and some people might not have it enabled, so it's maybe not an issue for them.

I thought of hiding that specific warning only on my end, but I believe tsserver is what's used for VSCode and other IDEs too; everyone using it might benefit from the change.

kkoomen:
> If it's not too much trouble, you could omit those type annotations only through a configuration variable, leaving the default as it is currently.

I think I should. Some prefer it, some don't.

Since we're doing quite a big rewrite at the moment (see #65) all features will be done after #65 has been merged.

PR #235, opened 2021-02-01
Add omit_redundant_param_types option to the javascript generator.

vantreeseba:

# Why this PR?

This would close issue #68.

It also makes the JSDoc work basically perfectly for haxe as well as typescript.


vantreeseba:
@kkoomen
Whenever you have some time, I couldn't get the tests to run locally (also I will make another commit with better naming / etc on the test), could you let me know if I am running the tests locally incorrectly?

I see there is a scripts/run-vader-tests.sh and I ran that as
```
./scripts/run-vader-tests.sh vim
```

But it just hung, is there something else I should be doing / need installed to so that I can attempt to clean this PR up and get it to the point where it is acceptable?

kkoomen:
@vantreeseba I don't think it's stuck. It just takes a damn long time to finish. Anyway, you can install the vim plugin Vader locally and then open the test file and run `:Vader` in there. That will be the quickest way to test your own file.

README.md
```diff
@@ -279,12 +279,16 @@ being aliased, see [`g:doge_filetype_aliases`](#gdoge_filetype_aliases).
...
 let g:doge_javascript_settings = {
 \  'destructuring_props': 1,
+\  'omit_redundant_param_types': 0,
 \}
 ```
 
 - `destructuring_props`: Whether or not to generate `@param` tags for the
   destructured properties in a function expression.
 
+- `omit_redundant_param_types`: Whether or not to omit the `{Type}` part of
```
kkoomen, line 289: - Please use `{type}` where the first letter is lowercase.

test/filetypes/javascript/functions.vader
```diff
@@ -373,6 +373,46 @@ Expect javascript (generated comments with a description, @param and @return tag
    */
   function foo({ a, b: str, c = 3, d: int = 5 }) {}
 
+# ==============================================================================
+# Test the 'omit_redundant_param_types' option
+# ==============================================================================
+Given javascript (functions with type information):
```
kkoomen, line 379: Please use the following description:

`functions where the type should be omitted if let g:doge_javascript_settings = {'omit_redundant_param_types': 1} is set`

to make it easier to find this test in the whole series of tests.

kkoomen (changes requested):

vantreeseba:
Both of the requested changes should be implemented, and I did get vader working locally.

Thanks so much for the help.

autoload/doge/preprocessors/javascript.vim
```diff
@@ -55,15 +55,19 @@ function! doge#preprocessors#javascript#tokens(tokens) abort
     if get(g:doge_javascript_settings, 'omit_redundant_param_types') == v:true
       let l:filtered_params = []
       for l:param in a:tokens['parameters']
-        if has_key(l:param, 'type') == v:true && !empty(l:param['type'])
-          let l:param['not_redundant'] = v:false
+        if has_key(l:param, 'type') == v:true
+          if(!empty(l:param['type']))
```
kkoomen, line 59: - Please add a space after `if`
- Remove the unnecessary parenthesis around `!empty()`

kkoomen (changes requested):

autoload/doge/preprocessors/javascript.vim
```diff
@@ -52,6 +52,25 @@ endfunction
 " function will adjust the input if needed.
 function! doge#preprocessors#javascript#tokens(tokens) abort
   if has_key(a:tokens, 'parameters') && !empty(a:tokens['parameters'])
+    if get(g:doge_javascript_settings, 'omit_redundant_param_types') == v:true
+      let l:filtered_params = []
```
kkoomen, line 56: You can just remove this variable, as you don't use it, right?

kkoomen (changes requested):

vantreeseba:
I realized the behavior of the return types is not correct, I will add another test and get some more code in to correct this.

I.e. Currently, no matter what the return is hidden (when omit_redundant_param_types is true), but it should only be hiding the type template portion is the type is known.

kkoomen:
@vantreeseba Regarding that the `./scripts/run-vader-scripts.sh` is not working locally is because in the CI it does clone the vader inside the current directory instead of being a sibling of the current working directory. I'll try to fix this asap.

kkoomen:
@vantreeseba Please fix tests so I can merge this soon. I want to merge this ASAP.

vantreeseba:
Hmm, I'm not sure why these tests are failing, I ran vader on those files locally and it seems to be working fine.  If you have any insight into why they would be, let me know and I will correct it.

kkoomen:
@vantreeseba The way Vader works when you pass multiple tests is that it will use a single vim instance to run everything. If tests fail, but locally running `:Vader` on the single file that fails, but doesn't fail locally, then the problem is that a combination of previous tests + current test will make the current test fail.

I'll make some time to fix the tests, because it's a bit troublesome to debug this.

VVKot:
Hey @kkoomen / @vantreeseba, wonder if there was any progress here? Anything I can help with?

kkoomen:
@VVKot I have time this weekend, so I can try to fix this.

kkoomen (approved):

kkoomen:
This feature has been merged and released in [v3.10.0](https://github.com/kkoomen/vim-doge/tree/v3.10.0).

PR #290, opened 2021-03-12
Add support for Rust

### LINKED ISSUE #13: Add support for Rust
I was actually kind of surprised to see that neither is Rust supported, nor is there an open issue for it. Though admittedly I haven't really looked for Rust documentation Vim plugins, this particular plugin seems quite nice.  
Rust has special documentation comments: `///` and `//!`. (It also supports markdown in its comments)

kkoomen:
This PR adds support for Rust with the following doc standards: [RustDoc](https://doc.rust-lang.org/book/ch14-02-publishing-to-crates-io.html)

Closes #13

# Features

- [x] Functions
  - [x] Implement `Arguments` section
  - [x] Implement `Errors` section
  - [x] Implement `Safety` section
  - [x] Implement `Examples` section

# Testing locally

For those who want to test this out locally, do a checkout to `feature/rust`, run `npm install` followed by `npm run build`.


arxra:
Works great for functions!  However, I don't have any `unsafe` code to test with unfortunatly.
Couldn't help but notice this only covers functions, not structs or enums which makes me question if it closes #13. It might be misleading to mark this as "supports rust" while the implementation covers functions

kkoomen:
@arxra No language I have supports everything and the main thing people add documentation for is functions for any language, only because it saves time to press a single button instead of writing it all by hand from scratch, that's why this plugin exists. I'm not gonna add support for things like `struct` if the only thing people add is a single comment, something like :

```
// Our main user struct used throughout the whole app.
struct User { }
```

If you have suggestions of templates, types of methods, variables, etc that could be of use and would safe people some time, feel free to comment it and we'll discuss it here.

arxra:
@kkoomen  that makes sense,  I just compared how it behaves in C and cpp where it does template for `structs` as I expected similar effects on different languages: 
```cpp
**
 * struct lims - [TODO:description]
 */
struct lims{
  int low;
  int high;
};
``

**Comment by kkoomen** [comment]
> @kkoomen that makes sense, I just compared how it behaves in C and cpp where it does template for `structs` as I expected similar effects on different languages (cpp down below):
> 
> ```c++
> **
>  * struct lims - [TODO:description]
>  */
> struct lims{
>   int low;
>   int high;
> };
> ```

It depends on the standards being used in the language. C-family languages have multiple standards compared to other languges and the one you mentioned is one of them, thereforce it's not relevant in this discussion. 

We're implementing support for Rust through rustdoc here, so we have to look at the way rustdoc is commenting things and also what the most common way is from the community. I personally have too less experience of Rust in order to properly judge this myself, hence I'm asking input from the folks in #13.

**Comment by arxra** [comment]
The documentation is indeed simpler here. I was probably expecting to be dropped in a
```
/// [TODO:description]
```
The `///` is a real minor nitpick from how I document with doge (spamming leader-d whenever I need a comment readable outside of the document) where this would come in handy.

All in All I take back what I said about not implementing #13, this covers all the nitty bits of writing documentation and will speed up the process for that since function parameters is the real killer. This may be a case of my specific usage of doge, which is outside the scope of this PR. If others feel like this is missing I'm sure you'll get a issue soon enough anyways 😅 (or a PR if lucky!)

File: README.md [hunk 1]
```diff
@@ -83,6 +83,7 @@ Is your favorite doc standard not supported?
 | :white_check_mark: | C++                                            | [Doxygen][doxygen]                                                           |
 | :white_check_mark: | C                                              | [Doxygen][doxygen], [KernelDoc][kerneldoc]                                   |
 | :white_check_mark: | Bash                                           | [Google][sh-google]                                                          |
+| :white_check_mark: | Rust                                           | [RustDoc][rustdoc]                                                              |
```
LanderMoerkerke, line 86: Some aligning

```suggestion
| :white_check_mark: | Rust                                           | [RustDoc][rustdoc]                                                           |
```

LanderMoerkerke (approved):

kkoomen:
This feature has been merged and released in [v3.9.0](https://github.com/kkoomen/vim-doge/tree/v3.9.0).

PR #297, opened 2021-06-07
Abide by numpy C0199 and W9011 specifications

### LINKED ISSUE #296: Generated Numpy API documentation out of sync with pylint's interpretation
**Describe the bug**

The generated Python docstring for a function, using Numpy style, provokes warnings from pylint. Specifically, from pylint's pylint.extensions.docparams plugin. It is a great plugin since it will notify me when the API docstring is no longer in sync with an evolving function signature.

See below for before & after examples, but in a nutshell, these two changes are needed:
1. add line continuation after the opening triple quote
2. remove the trailing colon on the type in the Returns section 

**Settings**

DoGE settings:

```vim
let g:doge_doc_standard_python = 'numpy'
```

Relevant pylint settings via `pylintrc` file:

```
load-plugins=pylint.extensions.docparams,pylint.extensions.docstyle,pylint.extensions.typing
```

**To Reproduce**

Two python functions below. Line numbers included to correlate with `pylint` error messages.

The first functions's API docstring (lines 29-42) is what DoGe generates. It provokes the pylint errors seen on lines 35 & 36.

The second function has two slight modifications on lines 47 & 56 -- see comments at lines 50-52 for details -- that keep pylint happy.

```

    28	def first_example() -> str:
    29	    """
    30	    Simplified function to demonstrate return code problem.
    31
    32	    This DoGe generated docstring will provoke the following warnings from
    33	    pylint with the pylint.extensions.docparams plugin loaded:
    34
    35	        demo.py:28:0: C0199: First line empty in function docstring (docstring-first-line-empty)
    36	        demo.py:28:0: W9011: Missing return documentation (missing-return-doc)
    37
    38	    Returns
    39	    -------
    40	    str:
    41	        standard greeting.
    42	    """
    43	    return "hello world"
    44
    45
    46	def second_example() -> str:
    47	    """\
    48	    Simplified function to demonstrate how to remove pylint errors.
    49
    50	    Adding the '\' character after the triple quotes above suppresses C0199.
    51
    52	    Removing the : after "str" below suppresses the W9011 warning.
    53
    54	    Returns
    55	    -------
    56	    str
    57	        standard greeting.
    58	    """
    59	    return "hello world"
```

**Expected behavior**

The Numpy specification, at least according to https://numpydoc.readthedocs.io/en/latest/format.html#returns, does not include a trailing ':' after the return type.

Taking pylint's warnings at face value, the first line of a docstring may not be empty and a bare """ on a line counts as empty.

kkoomen:
Fixes #296.

### DIFF
--- a/ftplugin/python.vim
+++ b/ftplugin/python.vim
@@ -102,7 +102,7 @@ call doge#buffer#register_doc_standard('numpy', [
 \      ],
 \    },
 \    'template': [
-\      '"""',
+\      '"""\',
 \      '!summary',
 \      '',
 \      '!description',
@@ -113,7 +113,7 @@ call doge#buffer#register_doc_standard('numpy', [
 \      '%(returnType|)%',
 \      '%(returnType|Returns)%',
 \      '%(returnType|-------)%',
-\      '%(returnType|{returnType}:)%',
+\      '%(returnType|{returnType})%',
 \      '%(returnType|\t!description)%',
 \      '%(exceptions|)%',
 \      '%(exceptions|Raises)%',
--- a/test/filetypes/python/functions-doc-numpy.vader
+++ b/test/filetypes/python/functions-doc-numpy.vader
@@ -12,7 +12,7 @@ Do (trigger doge):
 
 Expect python (generated comment with nothing but the summary and description 'TODO'):
   def myFunc():
-    """
+    """\
     [TODO:summary]
 
     [TODO:description]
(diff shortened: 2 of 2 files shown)


scottbolte:
Unfortunately while `pylint`'s docstring checks will accept a line continuation after the opening triple quote, `flake8` is pickier about leading white space. To make them both happy, the first line of text should be on the same line as the quotes.

scottbolte:
Also, I just found that `pylint` views Raises the same way as it does Returns. Neither should have a trailing colon after the exception/return type.

kkoomen:
> Unfortunately while `pylint`'s docstring checks will accept a line continuation after the opening triple quote, `flake8` is pickier about leading white space. To make them both happy, the first line of text should be on the same line as the quotes.

Does this mean the leading slash is also required? So, should it be:

```
def foo():
  """First line here.
  Second line here.
  """
```

or should it be like this?

```
def foo():
  """First line here.\
  Second line here.
  """
```

PR #367, opened 2022-06-04
Fix Expansion from the middle of multiline defs

janwaltl:
# Prelude

# Why this PR?

Expanding multi-line definitions only works correctly if the cursor is already on the first line. I work with C++ where definitions often span multiple lines due to templates or long return values and either I have to force myself to move the cursor to the first line or I get misplaced docstring.

Currently this

    template<typename T>
    long_return_type
    foo(); <--- Cursor

expands into

    template<typename T>
    long_return_type
    /* DOCSTRING */
    foo();

instead of this

    /* DOCSTRING */
    template<typename T>
    long_return_type
    foo();

So I made a small change which inserts the docstring on the first empty line instead of the fixed -1 offset.

I am not quite sure where should I make any tests for this, please let me know and I will try to add them.

### DIFF
--- a/autoload/doge/pattern.vim
+++ b/autoload/doge/pattern.vim
@@ -95,11 +95,11 @@ function! doge#pattern#generate(pattern) abort
     let l:comment_lnum_insert_position = line('.')
   else
     let l:comment_indent = 0
-	" Search for find empty (whitespace-only) line above.
-	let l:insert_l = line('.')
-	while (l:insert_l > 0) && getline(l:insert_l) !~ '^\s*$'
-		let l:insert_l -= 1
-	endw
+    " Search for find empty (whitespace-only) line above.
+    let l:insert_l = line('.')
+    while (l:insert_l > 0) && getline(l:insert_l) !~ '^\s*$'
+      let l:insert_l -= 1
+    endw
     let l:comment_lnum_insert_position = l:insert_l
   endif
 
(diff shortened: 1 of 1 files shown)


janwaltl:
Nevermind, I tried experimenting with this "fixed" code and it of course does not work if you have some text above the function, e.g. opening namespace braces or class definition. Not sure if we can get the first line of the definition in some other more reliable way.

kkoomen:
@janwaltl There are ways to fix this, because I also do this in other languages. I'll fix it ASAP.

kkoomen:
@janwaltl #375 fixes this problem, once the tests succeed, I'll do a new release. Thanks in advance for reporting this.

janwaltl:
@kkoomen  Thank you for fixing this! Yea, using parser for this is way better.

kkoomen:
This feature has been merged and released in [v3.13.0](https://github.com/kkoomen/vim-doge/tree/v3.13.0).

kkoomen:
@janwaltl I've adjusted the tests as well and noticed a small issue. The remaining fixes will be resolved once #378 is merged. I'll do a release for v3.13.1 today and will let you know.

kkoomen:
@janwaltl If this action has finished https://github.com/kkoomen/vim-doge/actions/runs/2514205500 then you can upgrade to the latest version, v3.14.0 :)

janwaltl:
Cool, thanks again very much!

PR #499, opened 2022-11-28
fix: avoid redefine command error

mars90226:
# Prelude

# Why this PR?

When upgrading pluging, the `plugin/doge.vim` will be executed twice (one for initial load of vim-doge, one for loading upgraded vim-doge). Using `:command` to redefine will cause vimscript error, which can be easily avoided by using `:command!`.


kkoomen (approved):
I agree. Thanks for noticing!

PR #618, opened 2023-09-10
fix(rust): handle destructured params case (#617)

### LINKED ISSUE #617: Rust: Function parameters as patterns causes panic in helper
**Describe the bug**
Consider the following function:
```rust
pub fn do_thing((i, j): (u8, u8)) -> u8 {
    i * j
}
```
Apparently, you can use irrefutable patterns in function parameters to deconstruct your input.

But the parser in the helper does not account for that, leading to a panic:
```bash
$ vim-doge-helper --filepath src/test.rs --parser rust --line 1 --doc-name rustdoc
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', src/rust/parser.rs:108:34
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
https://github.com/kkoomen/vim-doge/blob/4a1b44fa7c09556b79e87316ede5a7a45ff601dc/helper/src/rust/parser.rs#L103-L108

**To Reproduce**

1. Create a file containing the code above
2. Try to create documentation for the function `do_thing`

**Expected behavior**

**Output of `./bin/vim-doge-helper --version`**
```
vim-doge-helper 4.3.2
```
Built from source using commit 4a1b44fa7c09556b79e87316ede5a7a45ff601dc

**Output of `vim --version`**
Not applicable

kkoomen:
Fixes #617

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -15,7 +15,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest]
-        vim-version: [v7.4.2119, v8.2.5172, v9.0.150]
+        vim-version: [v7.4.2119, v8.2.5172, v9.0.1500]
     runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout kkoomen/vim-doge
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -15,7 +15,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest]
-        vim-version: [v7.4.2119, v8.2.5172, head]
+        vim-version: [v7.4.2119, v8.2.5172, v9.0.150]
     runs-on: ${{ matrix.os }}
     steps:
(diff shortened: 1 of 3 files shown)


RasmusRendal:
In some cases, it just seems to ignore arguments?
```rust
pub struct SorryForBeingMean {
    pub x: i8,
    pub y: i8,
}

pub fn some_function(SorryForBeingMean {x, y}: SorryForBeingMean) -> i8 {
    x * y
}
```

```rust
pub struct SorryForBeingMean (i8, i8);

pub fn some_function(SorryForBeingMean (x, y): SorryForBeingMean) -> i8 {
    x * y
}
```

You can do crazy things with these arguments, and I'm not even sure what the correct way to write rustdocs for these things are, so I totally understand if you don't want to support everything. For reference, this was the pattern I was trying to document: https://docs.rs/actix-web/latest/actix_web/web/struct.Form.html#examples

Thanks for the good work! :heart:

kkoomen:
Mhm, there doesn't seem to be anything specific indeed about how to write it. I'll just keep it like this for now and when rustdoc will be more mature, I'll just receive some PRs about this at some point in time :)

PR #640, opened 2023-10-27
Implement doxygen doc standard for python

### LINKED ISSUE #637: Consider adding doxygen for python
**Documentation standard**
According to e.g. https://www.woolseyworkshop.com/2020/06/25/documenting-python-programs-with-doxygen/ doxygen supports its syntax in python docstrings ( where applicable ).

It's not the most popular style, but in cases where you have multiple languages, something like doxygen that supports all of them might be useful ( we started using python in such an environment and currently experimenting with doxygen support, so I can't give much examples of how it would look like )

Maybe every supported language should have a doxygen compatible documentation standard available? 

**How popular is the doc standard?**
- [ ] Not popular at all
- [x] Some people use it
- [ ] Quite a lot of people use it
- [ ] It's not the most popular one, but a good alternative
- [ ] Everyone is using it

kkoomen:
Closes #637

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,24 +74,24 @@ Is your favorite language not supported?
 Is your favorite doc standard not supported?
 [Suggest a new doc standard][suggest-doc-standard] :tada:
 
-|                    | Language                                               | Doc standards                                                                |
-| :----------------- | :----------------------------------------------------- | :--------------------------------------------------------------------------- |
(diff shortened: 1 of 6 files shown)


kkoomen:
This feature has been merged and released in [v4.6.0](https://github.com/kkoomen/vim-doge/releases/tag/v4.6.0).

PR #670, opened 2025-02-05
Handle case when rust macro node has no children of type identifier

TabulateJarl8:
# Prelude

Thank you for helping out vim-doge!

By contributing to vim-doge you agree to the following statements:

# Why this PR?

To do an unhandled unwrap expression in [the rust parser](https://github.com/kkoomen/vim-doge/blob/750d17dcd5a38a6afb51ddd07ee7d3e96bab7e9e/helper/src/rust/parser.rs#L140), vim-doge would crash if a macro within a function had no children with the type of identifier (from my understanding of the code). A minimal reproducible example of this would be with the `log` crate:

 ```rs
fn testing() {
  log::info!("this will crash");
}
```

Note that this doesn't even have to be within a Rust project as it will crash either way. I fixed this bug by just saying that the item doesn't panic if it has no children which are identifiers, which I believe was the intended behavior.

 If you'd like me to remove the other changes that `rustfmt` automatically did when I saved the file, I can do that, but it did make it more consistent as there we're a few spots with 2-space indentation while the rest of the file was 4-space.

Thanks so much!

### DIFF
--- a/helper/src/rust/parser.rs
+++ b/helper/src/rust/parser.rs
@@ -1,5 +1,5 @@
-use tree_sitter::{Parser, Node};
 use serde_json::{Map, Value};
+use tree_sitter::{Node, Parser};
 
 use crate::base_parser::BaseParser;
 use crate::traverse;
@@ -36,7 +36,7 @@ impl<'a> BaseParser for RustParser<'a> {
                         new_line = sibling_node.start_position().row + 1;
                         temp_node = sibling_node;
                     } else {
-                      break;
+                        break;
                     }
                 }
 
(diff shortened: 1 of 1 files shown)


kkoomen (approved):

kkoomen:
@TabulateJarl8 please rebase from master again, because I had to update the workflow files to make the automated tests work again :)

TabulateJarl8:
@kkoomen should be done now. Not sure if it wants you to re-approve or if the CI just takes a minute to spin up, but seems like it wants approval on my end

